PR #97, opened 2017-09-13
Remove usage of C++14 feature 'return type deduction'.

maniac103:
The readme states the only used C++14 feature are binary literals, so
let's keep it like that if trivially possible like in this case.


maniac103:
Issue is similar to #62.

maniac103:
Oh, I see what I missed. I should actually look at the code instead of judging by compile result only. I'll need to dig deeper to come up with a more complete solution.

maniac103:
https://github.com/redboltz/mqtt_cpp/pull/98 should be better.

PR #204, opened 2019-04-04
test_broker.hpp use as::const_buffer instead of as::buffer

jonesmz:


jonesmz:
Never mind, a misunderstanding on how as::const_buffer is constructed.

PR #505, opened 2019-10-08
Use downloaded openssl instead of compiling ourselves

jonesmz:


jonesmz:
Never mind this one. I got the larger change working, which includes this.

PR #690, opened 2020-10-17
Number of performance optimizations in subscription map and retained_topic_map

kleunen:


test/subscription_map_broker.cpp
```diff
@@ -219,7 +219,7 @@ BOOST_AUTO_TEST_CASE( multi_non_wc_crud ) {
                     e.h = h.first; // update handle
                 }
             );
-            BOOST_TEST(!elem.h.empty());
+            BOOST_TEST(!!elem.h);
```
jonesmz, line 222: imho this isn't an improvement. The explicit .empty() call is more readable.

kleunen: it is because the type changed from vector to optional, the optional does not have an empty method

PR #702, opened 2020-10-21
Provide modify interface subscription map

kleunen:


kleunen:
I guess this modifies all matches and not a subscription. This is not complete i would say

PR #784, opened 2021-01-21
Example of a reconnect client

kleunen:
I added a no_tls_async_reconnect_client example which shows a "long-lived" client connection, which survives disconnects from the broker.

### DIFF
--- a/example/long_lived_client.cpp
+++ b/example/long_lived_client.cpp
@@ -1,0 +1,303 @@
+// Copyright Wouter van Kleunen 2020
+//
+// Distributed under the Boost Software License, Version 1.0.
+// (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+#include <iostream>
+#include <regex>
+
+#include <mqtt_client_cpp.hpp>
+
+#include <boost/uuid/uuid.hpp>
+#include <boost/uuid/uuid_generators.hpp>
+#include <boost/uuid/uuid_io.hpp>
+
+// This example shows the client reconnecting to the broker
+//
+// The client connects to the server and published a total of 100 messages,
+// if the connection was lost a new connection will be established.
+//
+// Important: please note that messages are only republished to the broker if
+// the broker still has an active session for this client. If it does not,
+// the client will start with a new session and not resend offline stored messages
+//
+// It is possible to connect to the server using (mqtt, mqtts, ws or wss) as follows:
+// long_lived mqtt://example.com
+// long_lived mqtts://example.com
+// long_lived ws://example.com
+// long_lived wss://example.com
+//
+// Server certificate is validated using the file cacert.pem
+
+
+template <typename C>
+void reconnect_client(boost::asio::steady_timer& timer, C& c)
+{
+    std::cout << "Start reconnect timer" << std::endl;
+
+    // Set an expiry time relative to now.
+    timer.expires_after(std::chrono::seconds(5));
+
+    timer.async_wait([&timer, &c](const boost::system::error_code& error) {
+        if (error != boost::asio::error::operation_aborted) {
+            std::cout << "Reconnect now !!" << std::endl;
+
+            // Connect
+            c->async_connect(
+                // [optional] checking underlying layer completion code
+                [&timer, &c]
+                (MQTT_NS::error_code ec){
+                    std::cout << "async_connect callback: " << ec.message() << std::endl;
+                    if (ec && ec != boost::asio::error::operation_aborted) {
+                        reconnect_client(timer, c);
+                    }
+                }
+            );
+        }
+    });
+}
+
+template <typename C>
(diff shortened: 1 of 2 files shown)


example/no_tls_async_reconnect_client.cpp
```diff
@@ -0,0 +1,195 @@
...
+{
+    // Publish a message every 5 seconds
+    timer.expires_from_now(boost::posix_time::seconds(5));
+
+    timer.async_wait([&timer, &c, packet_counter = packet_counter](const boost::system::error_code& error) {
+        if (error != boost::asio::error::operation_aborted) {
+            auto message = std::make_shared<std::string>("packet #" + std::to_string(packet_counter));
+            c->async_publish(
+                "mqtt_client_cpp/topic1",
+                *message,
```
redboltz, line 55:
```suggestion
                MQTT_NS::allocate_buffer("packet #" + std::to_string(packet_counter))
```

*1
I think that the following overload is used.
https://github.com/redboltz/mqtt_cpp/blob/master/include/mqtt/endpoint.hpp#L2444-L2511
And lifetime is automatically managerd.
https://github.com/redboltz/mqtt_cpp/blob/master/include/mqtt/endpoint.hpp#L2504-L2508

example/no_tls_async_reconnect_client.cpp
```diff
@@ -0,0 +1,195 @@
...
+                MQTT_NS::qos::exactly_once,
+                // [optional] checking async_publish completion code
+                [message = message]
```
redboltz, line 58:
```suggestion
                []
```

example/no_tls_async_reconnect_client.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+#include <iostream>
+
+#include <mqtt_client_cpp.hpp>
+
+// This example shows the client reconnecting to the broker
+//
+// Important: please note that messages are only republished to the broker if
+// the broker still has an active session for this client. Otherwise a clean
+// start is performed
```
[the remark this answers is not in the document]
kleunen: A clean session. It means the client will restart from message_id = 0 with sending, and not resend any stored offline messages.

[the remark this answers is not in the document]
redboltz: Clean Start is MQTT v5 and Clean Session is MQTT v3.1.1.
Clean Start does NOT mean session state expiration on the broker.
Clean Start is determined at the client side. Clean start is a request that cleaning existing session in the broker.
If the broker receive Clean Start : 1, then return Session Present: 0 in the CONNACK packet. When the client receives the CONNACK, then clear the client side session state that contains resend message.

Could you update the comment?

example/long_lived_client.cpp
```diff
@@ -0,0 +1,296 @@
...
+            // [optional] checking async_disconnect completion code
+            []
+            (MQTT_NS::error_code ec){
+                std::cout << "async_disconnect callback: " << ec.message() << std::endl;
+            }
+            );
+    };
+
+    // Setup client
+    c->set_client_id("reconnect_client");
```
kleunen, line 99: Generate an UUID here maybe ?

redboltz:
What is the purpose is the example?
I got confused. I don't understand the value of the example.

kleunen:
To show automatic reconnection, but I think it is also worthwhile to show how different client types should be handled (mqtt/mqtts/ws/wss). Because they don't derive from a base class, you have to use template functions to configure them.

redboltz:
If you want to demonstrate automatic reconnection, the example should be simpler.
The example is too much complicated. It looks like aging test tool. It is not suitable for the example.

kleunen:
I created new PR #785

with only reconnection

I think creating a client from an URL is interesting also, but maybe for another example

Radagan:
@kleunen, Thank you for providing this example. 
@redboltz, Respectfully as the learner, I must disagree. I feel strongly that this library, though very good, is difficult to use because it suffers from overly simplistic examples--code that would never be used in a real project. So, one is left cobbling these examples together to address real use cases such as reconnects. Who has a client device that only sends a few data items? It will most likely be a daemon process that has to stay running. Anyway, that has been my experience coming up to speed with mqtt_cpp.

redboltz:
@Radagan I updated tutorial.
https://github.com/redboltz/mqtt_cpp/wiki 
It is very welcome writing complex examples on your site. In mqtt_cpp's wiki, I would add the link to the examples.

Radagan:
Thank you for the reply. I would be happy to do so. In fact, I have planned a personal site for just that sort of thing. At this point though, I'm still working out what those best practices are with your library--that's the very reason those examples would be welcome! Catch-22. ;^) I'm using mqtt_cpp in an active work related project where those examples would be indispensable. Keeping the MQTT side running with 15 min reconnects for longer than a day is very problematic. So, my comments are not academic, but honest frustration with what seems to be a very promising library.

Sent with [ProtonMail](https://protonmail.com) Secure Email.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

PR #850, opened 2021-09-03
Removed unused handler.

redboltz:
Fixed test precondition to increase coverage.


redboltz:
Coverage improvement observed as I expected.

PR #1000, opened 2025-07-21
Sonatus master

redboltz:


arevetchi:
@redboltz can we have a new version released? or is one planned for near future?

redboltz:
@arevetchi I will release the new version ASAP.

